PR #8, opened 2015-06-01
replace all instances of "-" and "_" in the token instead of the first one

ivan-st:


jfromaniello:
Thank you very much, published as `v1.1.1`

PR #10, opened 2015-06-05
Handle undefined token gracefully

knownasilya:
If there is no token, then this function errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ console.log(decoded);
  */
 ~~~
 
+**Note:** A falsy token will throw an error.
+
 ## Develop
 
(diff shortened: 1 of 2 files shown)


knownasilya:
Ping :+1:

dschenkelman:
Thanks for the PR @knownasilya! I don't think returning is a good idea. I would probably throw an error in this case, as it is an invalid argument. Would you mind making the update to error with a clearly defined message and also document that?

Thanks!

knownasilya:
@dschenkelman is that good what you had in mind?

jfromaniello:
It looks good to me, I released it as `v1.2.0`.

knownasilya:
Thanks.

PR #11, opened 2015-07-07
Avoid use of window in parsing JSON

### LINKED ISSUE #5: window is not always available
The [method for parsing json](https://github.com/auth0/jwt-decode/blob/master/lib/json_parse.js#L2) depends on `window` being available which prohibits use in node.js

tschaub:
This uses `JSON.parse()` if `JSON` is defined.  Otherwise, it uses `eval`.

Fixes #5.

### DIFF
--- a/lib/json_parse.js
+++ b/lib/json_parse.js
@@ -1,3 +1,9 @@
 module.exports = function (str) {
-  return window.JSON ? window.JSON.parse(str) : eval('(' + str + ')');
+  var parsed;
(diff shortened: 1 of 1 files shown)


tschaub:
It looks like the tests require more than `npm install && npm test` to run.

jfromaniello:
Thank you very much and sorry for the delay. All tests are green.

Released as `v1.1.2`.

PR #18, opened 2015-11-18
README info on Polymer wrapper

mkazlauskas:
I created a Polymer-based web component wrapper. Just another option to install and use it.


mkazlauskas:
Added entry-point file for HTML Imports, which ensures that it's only loaded once when using with web components.

jfromaniello:
Thanks a lot and sorry for the delay

PR #19, opened 2015-11-22
check for string type

mattkrick:
gotta make sure there's a split method on the token, otherwise it throws.


jfromaniello:
thanks!

Issue #24, opened 2016-02-19
Remove eval function


euharrison:
Hi! I received an alert from my host that the script has a suspicious `eval()` code, this is the alert:

![12637151_955493134530531_662725185_o](https://cloud.githubusercontent.com/assets/557598/13178500/4bf57868-d706-11e5-8743-bf49c55f5a8b.jpg)

Debugging it I found that this code came from here https://github.com/auth0/jwt-decode/blob/master/build/jwt-decode.js#L56

Do you think that we have a workaround with this eval function?

jfromaniello:
It is using eval as a fallback when there is no JSON support in the browser... maybe we can remove that since it seems is already supported in every major browser:

http://caniuse.com/#feat=json

and if the developer really needs to support old browser he can polyfill on their own before loading this library.

sandstrom:
@jfromaniello sounds like a great solution! JSON is *widely* supported, and there are good polyfills available for those rare, old browsers.

jfromaniello:
thanks!

PR #25, opened 2016-03-05
remove eval function for json parse

euharrison:
https://github.com/auth0/jwt-decode/issues/24


euharrison:
This PR was exchanged by https://github.com/auth0/jwt-decode/pull/26 because it has the built files

Issue #33, opened 2016-06-24
receiving header instead of payload information


quaqua:
I'm in need to receive not the payload but the HEADER of a jwt. I was surprised when I found out, that jwt_decode just statically uses the [1] index of the jwt.
(here: https://github.com/auth0/jwt-decode/blob/master/lib/index.js#L10)

May I suggest to add an alternative option argument to handle that issue? Something like:

```javascript
jwt_decode(token,{ header: true });
```

or am I missing something fundamental?

PR #34, opened 2016-06-24
adds header option to jwt_decode

quaqua:
adds header option and proofs by test

affects #33


jfromaniello:
Thank you very much and sorry for the delay. Released as `v2.1.0`

Issue #28, opened 2016-05-02
Deprecation warnings for sourceURL and sourceMappingURL


boldt:
```
'//@ sourceURL' and '//@ sourceMappingURL' are deprecated, please use '//# sourceURL=' and '//# sourceMappingURL=' instead.
```
Version: 2.0.1

skattabrain:
+1

jfromaniello:
I fixed this upgrading browserify.

raymanoz:
Awesome, thanks!

PR #35, opened 2016-07-05
Eliminate warning about use //# instead of //@

raymanoz:
Attempting to fix issue #28


jfromaniello:
This file is generated automatically by browserify.. I will try to fix this issue by upgrading it.

jfromaniello:
Done in `v2.0.4` . Thanks and sorry for the delay

Issue #29, opened 2016-05-13
IE 11: Invalid calling object


brianleipprandt:
I'm seeing a 'Invalid calling object' exception in IE 11 when calling jwtDecode with my token.  Other browsers seem to all be working.

It looks like this may be related to not binding window.atob in lib/atob.js when the polyfill isn't used.

Stack:

```
{exception} :
  description: "Invalid calling object",
  message: "Invalid calling object",
  name: "TypeError",
  number: -2147418113,
  stack	"TypeError: Invalid calling object\n   at b64DecodeUnicode (eval code:4:3)\n   at module.exports (eval code:29:5)\n   at module.exports (eval code:10:3)\n   at updateAuthToken (eval code:34:3)\n   at auth (eval code:96:7)\n   at combination (eval code:117:7)\n   at dispatch (eval code:179:7)\n   at Anonymous function (eval code:208:13)\n   at Anonymous function (eval code:14:9)\n   at Anonymous function (eval code:140:9)"
```

Manual checks via debugger in b64DecodeUnicode:
```
window.atob('test')
"µë-"
atob('test')
Invalid calling object
atob
function atob() { [native code] }
atob === window.atob
True
var bound_atob = window.atob.bind(window);
undefined
bound_atob('test')
"µë-"
```

mikecousins:
As a heads-up, I couldn't wait any longer on this being fixed so I forked it and made a new version with this fix in.

https://www.npmjs.com/package/jwt-decode-ie

PR #31, opened 2016-06-21
(fixes #29) IE: Invalid calling object

bmatcuk:
Thanks to @brianleipprandt for figuring out the problem. This pull request is a slight improvement that fixes issues with browsers that do not have atob (such as IE9).

PR #37, opened 2016-07-14
use atob from abab npm module

pklingem:
fixes https://github.com/auth0/jwt-decode/issues/29


jfromaniello:
I have merged #31 . I think it the library is smaller as it is.

Sorry for the delay

PR #38, opened 2016-08-05
Update bower.json

sdoxsee:
The current release is 2.1.0. package.json is up-to-date. I think bower.json should be as well.


jfromaniello:
thanks

Issue #36, opened 2016-07-05
include type definition


PatrickJS:
can you include a type definition file? probably starting off with something like
`lib/index.d.ts`

``` typescript
function jwtDecode(token: string): any;
namespace jwtDecode {}
export = jwtDecode;
```

`package.json`

``` json
{
  "typings": "lib/index.d.ts"
}
```

related https://github.com/AngularClass/angular2-webpack-starter/issues/755#issuecomment-230537444

mbeauchamp7:
```typings search jwt-decode``` It's there

**Comment by jfromaniello** [comment]
I am willing to do this but I am not familiar with those tools, can you send me a PR?

(sorry for the delay)

**Comment by hanssens** [comment]
As also mentioned in the PR #42, I tend to agree with @mbeauchamp7 here; the typings are already available through:

```
# see also https://www.npmjs.com/package/@types/jwt-decode
npm install --save-dev @types/jwt-decode
```

Having it as part of the jwt-decode repo makes sense only if it is actively supported. Otherwise, I'd gladly trust the DefinitelyTyped version.

PR #42, opened 2016-11-09
Add TypeScript definition file

frontweb:
This PR adds TypeScript definition file and updates the README with an example of how to import the function in TypeScript.

Related to https://github.com/auth0/jwt-decode/issues/36

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,2 @@
+declare function jwtDecode(token: string): any;
+export default jwtDecode;
--- a/README.md
+++ b/README.md
@@ -31,6 +31,12 @@ Can also be used with [browserify] or [webpack] by doing `npm install jwt-decode
 var jwtDecode = require('jwt-decode');
 ~~~
 
+Can also be used with Typescript:
+
+~~~javascript
+import jwtDecode from 'jwt-decode';
+~~~
+
 ## Polymer Web Component
 
 Can also be installed and used with [Polymer-based wrapper](https://github.com/firmfirm/f-jwt-decode).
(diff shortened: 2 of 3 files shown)


hanssens:
@frontweb Out of curiosity; any particular reason why manually adding it through @types wouldn't suffice?

```
# see also https://www.npmjs.com/package/@types/jwt-decode
npm install --save-dev @types/jwt-decode
```

Don't get me wrong, being a consumer and TypeScript user I applaud the initiative, but having the .d.ts seperately seems to make more sense in the case where the repo-maintainers don't  intend to update it themselves. 

Also, the [DefinitelyTyped repo for jwt-decode.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/types-2.0/jwt-decode) seems to be properly in order, right?

frontweb:
@hanssens my understanding is that bundling the type definitions with the npm package is the preferred method lately.

lib/index.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function jwtDecode(token: string): any;
```
rebirthtobi: this doesn't support the support the second paranmeter `options`

```
    declare module "jwt-decode" {
        interface OptionsType {
            header: boolean;
        }
        type jwtDecodeType = (token: string, options?: OptionsType) => any;
        const jwtDecode: jwtDecodeType;
        export default jwtDecode;
    }
```

Can be modified though

frontweb:
Closing this since types are available through @types/jwt-decode

FredericLatour:
@frontweb @hanssens 

Bundling the type definitions has several advantages:
* less moving parts (you don't need a second package just for the types)
* it makes it more straightforward to keep types in sync.

Even if the maintainers do not write/update types themselves, it is always possible to submit a PR.

PR #48, opened 2017-03-17
Throwing a specific error

fiddur:
In order to use this in a larger `try / catch` block, we need to have a specific error type from this, not just en `Error` without a name.


jfromaniello:
Thank you, released as v2.2.0

PR #56, opened 2017-09-27
Perform an AMD check that is compatible with RequireJS' namespace feature

simong:
RequireJS its optimizer has [a feature to namespace all global variables](http://requirejs.org/docs/faq-optimization.html#namespace) such as `define`, `require` and `requirejs` with a namespace object. It can do this when the `define` check is done through `typeof define === 'function' && define.amd`.

Currently, an explicit check for global.window.define is done. This PR changes the behavior in standalone.js to achieve compatibility with RequireJS' optimizer.


simong:
@jfromaniello Any thoughts? Happy to change anything :)

PR #68, opened 2018-05-04
Support ES6 modules import

alexey-pelykh:


moderndeveloperllc:
@alexey-pelykh There are still some `require()` calls in there that are node.js specific, so you can't really use this change in an ES6 module-only environment without a build process.

alexey-pelykh:
Well, this change worked for me specifically for ES6, yet I'm working in node.js environment, just with TS.

PR #99, opened 2020-10-01
fix readme

Zelig880:

### Description

Our cypress tests broke after upgrading JWT to use the newest version (3.0). 
The issue was to do with the "changes" in the library is imported.

I have changed the Readme, so that if people look for example of using this library with Require, they will find the correct documentation

### References

https://stackoverflow.com/questions/43247696/javascript-require-vs-require-default

### Testing

i have tested out application after the change both manually and by running E2E with cypress and everything worked after applying the change above

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 
 **Warning: When upgrading from version `2` to `3`, there's a potentially breaking change**
 
+
 If you've previously imported the library as `import * as jwt_decode from 'jwt-decode'`, you'll have to change your import to `import jwt_decode from 'jwt-decode';`.
 
 ---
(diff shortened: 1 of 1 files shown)


Sambego:
Thanks for this pull request. 

The issue you were facing is because cypress uses the new ESM build, but you're importing the library with a CommonJS require. By default cypress will use that build as it can handle native JS imports.

Changing `const jwt_decode = require('jwt-decode');` to `import jwt_decode from 'jwt-decode';` will be the best solution here.

Should you use the CommonJS build, () the `require()` statement still works like before. Here's a small demo using Node:
https://codesandbox.io/s/gifted-gauss-rdvwo?file=/src/index.js

Issue #103, opened 2020-10-26
Build output inconsistency makes it impossible to create TypeScript definitions


remcohaszing:
### Description

This is basically DefinitelyTyped/DefinitelyTyped#49112, but as an issue in this repository.

Because the `main` entry uses `module.exports = function jwtDecode() {}` to export the function, and the `module` entry uses `export default function jwtDecode() {}`, it’s impossible to create TypeScript definitions for version 3 of this library.

The CommonJS export should be typed as:

```ts
declare function jwtDecode();
export = jwtDecode;
```

The ESM export should be typed as:

```ts
export default function jwtDecode();
```

This has caused various issues reported here for people who don’t know how to use this library.

#90 #91 (comments) #96 #97 #102

#### Solution A: Omit ESM module

Removing the ESM module and `module` entry from `package.json` would leave the CJS export. This deoptimizes module bundling, but looks the prettiest in a NodeJS environment.

NodeJS usage:

```js
const jwtDecode = require('jwt-decode');
```

Module / TypeScript usage:

```ts
import * as jwtDecode from 'jwt-decode';
```

Type definition:

```ts
declare function jwtDecode();
export = jwtDecode;
```

#### Solution B: Use default key in CJS

Another solution is to favor ESM syntax and use `module.exports.default` for CJS.

NodeJS usage:

```js
const { default: jwtDecode } = require('jwt-decode');
```

Module / TypeScript usage:

```ts
import jwtDecode from 'jwt-decode';
```

Type Definition:

```ts
export default function jwtDecode();
```

#### Solution C: Use named exports (My personal favourite)

Named exports look good in both CJS and ESM.

NodeJS usage:

```js
const { jwtDecode } = require('jwt-decode');
```

Module / TypeScript usage:

```ts
import { jwtDecode } from 'jwt-decode';
```

Type Definition:

```ts
export function jwtDecode();
```

### Reproduction

N/A

### Environment

> Please provide the following:

- **Version of this library used:** 3.0.0
- **Version of the platform or framework used, if applicable:**
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:** TypeScript, `@types/jwt-decode`

Blackglade:
Can confirm, just ran into this exact problem. What I am having a hard time understanding is why there are two different jwtDecode() functions exported?

remcohaszing:
#### Solution D: Assign `exports.default` in the `cjs` output.

[...]

PR #105, opened 2020-10-31
Add type definitions

remcohaszing:
### Description

This adds TypeScript type definitions based on #104.

If this isn’t desired, I’ll gladly add these to DefinitelyTyped instead once #104 is merged and released.

### References

#103
#104

### Testing

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,7 @@
+export class InvalidTokenError extends Error {}
+
+export interface JwtDecodeOptions {
+  header?: boolean;
+}
+
+export default function jwtDecode(token: string, options?: JwtDecodeOptions): unknown;
--- a/lib/index.cjs.js
+++ b/lib/index.cjs.js
@@ -1,5 +1,6 @@
 import jwtDecode, { InvalidTokenError } from "./index";
 
 const wrapper = jwtDecode;
+wrapper.default = jwtDecode;
 wrapper.InvalidTokenError = InvalidTokenError;
 export default wrapper;
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "description": "Decode JWT tokens, mostly useful for browser applications.",
     "main": "build/jwt-decode.cjs.js",
     "module": "build/jwt-decode.esm.js",
+    "types": "index.d.ts",
     "keywords": [
         "jwt",
         "browser"


glennreyes:
I saw this after I have written out myself types for that, maybe you might want to adapt:

```ts
declare module 'jwt-decode' {
  type JWTAlgorithm =
    | 'HS256'
    | 'HS384'
    | 'HS512'
    | 'RS384'
    | 'RS512'
    | 'ES256'
    | 'ES384'
    | 'ES512'
    | 'PS256'
    | 'PS384'
    | 'PS512'
    | 'none';

  type JWTType = 'JWT' | 'JWE';

  function jwtDecode<Payload>(
    token: string,
  ): Payload & { ext: number; iat: number };
  function jwtDecode<Payload>(
    token: string,
    options: { header: false },
  ): Payload & { ext: number; iat: number };
  function jwtDecode<Payload>(
    token: string,
    options: { header: true },
  ): { typ: JWTType; alg: JWTAlgorithm };

  export default jwtDecode;
}
```

remcohaszing:
@glennreyes Thanks for your suggestion!

The types you proposed are probably correct 99% of the time (if #104 gets merged). Unfortunately that doesn’t mean they’re always correct.

From https://jwt.io/introduction

> The header _typically_ consists of two parts: the type of the token, which is JWT, and the signing algorithm being used, such as HMAC SHA256 or RSA.

This means it could be something else, i.e. `unknown`.

The same reasoning goes for the payload. You picked `exp` and `iat`, but there are more registered claims, and all claims are optional. Also one could add an entirely custom payload.

If the maintainers decide to include my types **and** they decide it’s a good idea to provide types for the typical payload options, I’ll gladly include them based on your initial work. :smiley:

glennreyes:
> @glennreyes Thanks for your suggestion!
> 
> The types you proposed are probably correct 99% of the time (if #104 gets merged). Unfortunately that doesn’t mean they’re always correct.
> 
> From https://jwt.io/introduction
> 
> > The header _typically_ consists of two parts: the type of the token, which is JWT, and the signing algorithm being used, such as HMAC SHA256 or RSA.
> 
> This means it could be something else, i.e. `unknown`.
> 
> The same reasoning goes for the payload. You picked `exp` and `iat`, but there are more registered claims, and all claims are optional. Also one could add an entirely custom payload.
> 
> If the maintainers decide to include my types **and** they decide it’s a good idea to provide types for the typical payload options, I’ll gladly include them based on your initial work. 😃

Thank you for looking through it!

From the [spec], it can be indeed an arbitrary JSON object:


> JWTs represent a set of claims as a JSON object that is encoded in a
>   JWS and/or JWE structure.  This JSON object is the JWT Claims Set.
>   As per Section 4 of RFC 7159 [RFC7159], the **JSON object consists of
>   zero or more name/value pairs** (or members), where the names are
>   strings and the values are arbitrary JSON values

Therefore, it will make total sense to type it as `unknown` by default.
Providing the typical claims separately would be super handy!

Sambego (dismissed):

Sambego (approved):

Sambego:
Thanks for this pull request @remcohaszing, the changes will be available in version `v.3.1.0`

rm-hull:
@Sambego The use of the generic in the type definition from @types/jwt-decode is a bit nicer than returning unknown, see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jwt-decode/index.d.ts#L12 .. is that something that can be incorporated in a new patch?

Sambego:
I must admit that I'm far from an expert when it comes to type definitions. 
If using a generic is indeed a better solution, of course, I can create a patch for this. 

@remcohaszing is there any reason you chose `unknown` over the generic used in the type definition on definitely typed?

rm-hull:
It lets me declare upfront what my token payload contains, and would do the cast automatically (obvs. there is no runtime guarantee that it is the right shape):

```typescript
type JwtToken = {
  username: string;
  iat: number;
  exp: number;
  roles: string[];
};

const username = jwt_decode<JwtToken>(token).username
```
When it is unknown, this would have to be an explicit cast instead:

```typescript
const username = (jwt_decode(token) as JwtToken).username
```

remcohaszing:
The generic in the old type definition was unnecessary. On DefinitelyTyped this is considered a bad practice and reported by their linter. Because they maintain so many types, they disable some rules for each package specifically, as can be seen [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6970a8fffa0743f0f5fc918e187fa37f0d2675df/types/jwt-decode/tslint.json).

I believe using generics has a performance impact, although I think it’s tiny and I don’t know the details about this.

The usage based on the code above with current types would be:

```ts
const { username } = jwtDecode(token) as JwtToken;
```

I have no strong opinion on including or omitting the generic. I just tried to adhere to rules on DefinitelyTyped. :)

PR #107, opened 2020-11-05
add reserved claims types

dschaller:

### Description

> Describe the purpose of this PR along with any background information and the impacts of the proposed change. For the benefit of the community, please do not assume prior context.
>
> Provide details that support your chosen implementation, including: breaking changes, alternatives considered, changes to the API, etc.
>
> If the UI is being changed, please provide screenshots.

Add type definitions for reserved claims as listed in [section 4.1](https://tools.ietf.org/html/rfc7519#section-4.1) of the RFC.

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,12 @@ export interface JwtDecodeOptions {
   header?: boolean;
 }
 
-export interface JwtClaims {
+export interface JwtHeader {
+  type?: string;
+  alg?: string;
+}
+
+export interface JwtPayload {
   iss?: string;
   sub?: string;
   aud?: string[] | string;
@@ -14,4 +19,7 @@ export interface JwtClaims {
   jti?: string;
 }
 
-export default function jwtDecode<ReturnType = unknown>(token: string, options?: JwtDecodeOptions): ReturnType;
+export default function jwtDecode<T = unknown>(
+  token: string,
+  options?: JwtDecodeOptions
+): T;
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,4 +4,14 @@ export interface JwtDecodeOptions {
   header?: boolean;
 }
 
-export default function jwtDecode(token: string, options?: JwtDecodeOptions): unknown;
+export interface JwtClaims {
+  iss?: string;
+  sub?: string;
+  aud?: string[] | string;
+  exp?: number;
+  nbf?: number;
+  iat?: number;
+  jti?: string;
+}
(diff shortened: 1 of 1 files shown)


Sambego:
We're having a bit of a discussion about whether to return `unkown`, or provide the opportunity to use a generic to cast the returned type in the comments of #105.

I saw you have an internal pr that is blocked by this, but I would like to keep these types as flexible as possible. Would the generic approach where you can do the following also work for you?

```ts
type JwtPayload = {
  iat: number;
  exp: number;
  ...
  roles: string[];
};

const { username } = jwt_decode<JwtPayload>(token);
```

dschaller:
@Sambego that would work but I'd be curious why use unknown if the RFC defines reserved optional claims? Users could then extend the type interface if they have custom claims that they want to add.

felixmosh:
You can combine both approaches, generics with default value.

```js
export interface JwtClaims {
  iss?: string;
  sub?: string;
  aud?: string[] | string;
  exp?: number;
  nbf?: number;
  iat?: number;
  jti?: string;
}

export default function jwtDecode<T = JwtClaims>(token: string, options?: JwtDecodeOptions): T;
// or using extention
export default function jwtDecode<T extends JwtClaims = JwtClaims>(token: string, options?: JwtDecodeOptions): T;
```

This way, you can pass from the outside your type but by default it will be `JwsClaims`

index.d.ts
```diff
@@ -4,4 +4,14 @@ export interface JwtDecodeOptions {
...
+  iss?: string;
+  sub?: string;
+  aud?: string[] | string;
+  exp?: number;
+  nbf?: number;
+  iat?: number;
+  jti?: string;
+}
+
+export default function jwtDecode(token: string, options?: JwtDecodeOptions): JwtClaims;
```
vazgabriel, line 17: I'd say this is a really easy way to make it generic without changing too much
```suggestion
export default function jwtDecode<ReturnType = unknown>(token: string, options?: JwtDecodeOptions): ReturnType;
```

Sambego (approved):

Sambego:
Thanks, everybody for chiming in on this, your insights are appreciated 💪 

- Went with the suggestion by @vazgabriel as it would not introduce a breaking change
- Export a default type for a JWT header and payload, which can be used as-is, or extended, and passed to the generic.

Sambego:
These updates are available in `v3.1.2`

PR #118, opened 2021-08-24
make token parameter type unknown

SkipKayhil:
### Description

The jwtDecode function will throw an `InvalidTokenError` when the token
parameter is not a string, so since this is part of the api, the type of
the token should not be limited to just string.

Let me know if this seems reasonable or if there is anything I should change

### Checklist


Sambego:
Thanks for this Pull Request! This library only decodes JSON Web Tokens serialized as a string, setting this to unknown would enlarge the surface for errors.

PR #146, opened 2023-06-20
Update atob polyfill export mechanism

cristobal:
Change export mechanism.
1. Check first for presence of globalThis + atob.
2. Secondly check for window + atob
3. Finally polyfill if none of the above implementations are present

### Description
This ensures that we don't polyfill unecessary on the server side, since the expression `type of window !== "undefined"`, will always evaluate to false on the server side.

However by adding a check on [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) first we ensure to check on the globalThis object instead which is present both on the server and client.
Screenshot 2023-06-20 at 18 13 06

### References

There are no proposals for this, but it's a nice to have rather than using a polyfill on the server which is most likely not used.

This is also how it's done in [core-js](https://github.com/zloirock/core-js/blob/55608b6965cd445eb5588188b6be9f202a2d29bf/packages/core-js/internals/global.js#L5-L9):
```js
// https://github.com/zloirock/core-js/issues/86#issuecomment-115759028
module.exports =
  // eslint-disable-next-line es/no-global-this -- safe
  check(typeof globalThis == 'object' && globalThis) ||
  check(typeof window == 'object' && window) ||
```

**See support for GlobalThis**
* https://caniuse.com/?search=atob
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

**Articles supporting use of globalThis**
* [What is globalThis, and why should you start using it?](https://blog.logrocket.com/what-is-globalthis-why-use-it/)
* [WHY YOU SHOULD BE USING GLOBALTHIS INSTEAD OF WINDOW IN YOUR JAVASCRIPT CODE](https://ilikekillnerds.com/2023/02/why-you-should-be-using-globalthis-instead-of-window-in-your-javascript-code/)
### Testing

The current change should not affect testing, if there is a need for test i can add it. Which could check that the polyfill method is never called on the server. However not sure how much a need there is for this.

### Checklist

### DIFF
--- a/lib/atob.js
+++ b/lib/atob.js
@@ -39,7 +39,15 @@ function polyfill(input) {
     return output;
 }
 
-export default (typeof window !== "undefined" &&
-    window.atob &&
-    window.atob.bind(window)) ||
-polyfill;
+// globalThis supported and atob implemented
+if (typeof globalThis !== "undefined" && globalThis.atob) {
+    export globalThis.atob.bind(globalThis);
+} 
+// no globalThis support, however window present and atob implemented
+else if (typeof window !== "undefined" && window.atob) {
+    export window.atob.bind(window);
+} 
+// missing atob support, polyfill it.
+else {
+    export polyfill;
+}


lib/atob.js
```diff
@@ -39,7 +39,15 @@ function polyfill(input) {
...
-polyfill;
\ No newline at end of file
+// globalThis supported and atob implemented
+if (typeof globalThis !== "undefined" && globalThis.atob) {
+    export globalThis.atob.bind(globalThis);
+} 
+// no globalThis support, however window present and atob implemented
+else if (typeof window !== "undefined" && window.atob) {
+    export window.atob.bind(window);
+}
```
cristobal, line 49: Perhaps we could remove the whole check on `window`, and just assume `globalThis` only?

cristobal:
Gonna make another PR here closing this one.

PR #176, opened 2023-07-31
Clean up Rollup config

jonkoops:
Cleans up the Rollup config a little, by:

- Using a variable to store the `input` field, which is now the same everywhere
- Removing `[]` for output and using a single definition
- Removing `name` for outputs that don't need it
- Removing the explicit `exports` option for the output, as the default already matches.


jonkoops:
Added a commit that fixes the tests as well.

frederikprijck (approved):

PR #178, opened 2023-07-31
Explicit type import and exports (BETA)

cristobal:
Be explicit about that we are importing types only and what we are actually re-exporting.

Using a global `*` export can often be a pitfall in the long run because it assumes everything is export and often one will create some internal types that often were not intended for everyone to be exported also.

It's better to be clear that we are import types, and also what is meant for public consumption and not.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,8 +1,8 @@
 "use strict";
 
 import base64_url_decode from "./base64_url_decode";
-import { JwtDecodeOptions, JwtHeader, JwtPayload } from "./global";
-export * from './global';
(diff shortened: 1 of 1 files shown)


lib/index.ts
jonkoops, line 1: Personally, I'd rather see the types moved into this file. They aren't very large, and won't clutter things up too much, and all the types are relevant for the code in this file.

cristobal: I agree i can update the PR here 👋

frederikprijck (approved):

PR #182, opened 2023-07-31
Update docs to use named exports

jonkoops:
Reworks the docs so that it reflects the usage of named exports over default.


README.md
jonkoops, line 1: This could serve as reference for the migration guide for v4.

frederikprijck (approved):

PR #184, opened 2023-08-01
Remove `'use strict'` from source

jonkoops:
Removes the `'use strict'` statement from the source code. This is no longer required as modules are [strict by default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_modules), and Rollup will therefore automatically add `'use strict'` to the CommonJS and UMD bundles.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,5 +1,3 @@
-"use strict";
-
 import base64_url_decode from "./base64_url_decode";
 import { JwtDecodeOptions, JwtHeader, JwtPayload } from "./global";
 export * from './global';


cristobal:
Perhaps one should add `strict` to the tsconfig in this PR also?
Such that TypeScript also operates under the assuption of strict mode.

https://www.typescriptlang.org/tsconfig#strict

cristobal:
> TypeScript strict more dictates the strictness of types, it has nothing to do with strict mode in JavaScript.

Yes it turns on the [`alwaysStrict`](https://www.typescriptlang.org/tsconfig#alwaysStrict) flag in addition to more type checking strictness as you say.

> Ensures that your files are parsed in the ECMAScript strict mode, and emit “use strict” for each source file.

cristobal:
From [StackOverflow](https://stackoverflow.com/questions/43727889/is-it-a-good-idea-to-always-set-alwaysstrict-to-true-in-tsconfig-json#:~:text=The%20strict%20TypeScript%20flag%20is,use%20strict'%20in%20the%20output.) Strict vs alwasStrict

>It is a good idea to not only include alwaysStrict, but to include the strict flag that enables alwaysStrict and also noImplicitAny, noImplicitThis, strictBindCallApply, strictNullChecks, strictFunctionTypes, strictPropertyInitialization and useUnknownInCatchVariables which are even more important.

cristobal:
Does not matter seems that strict is enabled by default either way:
Screenshot 2023-08-01 at 13 11 11

Personally i would prefer to be explicit about it being enabled imo.

cristobal:
Seem that it is already enabled 😂 https://github.com/auth0/jwt-decode/blob/beta/tsconfig.json#L13

PR #187, opened 2023-08-05
Use modern JavaScript syntax

jonkoops:
Use more modern JavaScript syntax where possible, namely:

- [Arrow function expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) over [function expressions](https://developer.mozilla.org/en-US/docs/web/JavaScript/Reference/Operators/function).
- [Template literal strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) over [string concatenation](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Strings#concatenation_using).
- [Logical OR assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment) over assignment with Logical OR fallback.
- `let` of `const` over `var`.

### DIFF
--- a/lib/base64-url-decode.ts
+++ b/lib/base64-url-decode.ts
@@ -1,6 +1,6 @@
 function b64DecodeUnicode(str: string) {
   return decodeURIComponent(
-    atob(str).replace(/(.)/g, function(m, p) {
+    atob(str).replace(/(.)/g, (m, p) => {
       let code = p.charCodeAt(0).toString(16).toUpperCase();
       if (code.length < 2) {
         code = "0" + code;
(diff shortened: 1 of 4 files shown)


lib/base64-url-decode.ts
jonkoops, line 1: Took the liberty of changing the name of this file so the separators match the other files in the project.

lib/index.ts
```diff
@@ -20,11 +20,7 @@ export interface JwtPayload {
   jti?: string;
 }
 
-export class InvalidTokenError extends Error {
-  constructor(message: string) {
```
jonkoops, line 24: Since this constructor only passes the same arguments as the class it extends, I've removed it so that it will fall back to the parent constructor.

test/tests.ts
jonkoops, line 1: I've also changed the names of the variables here from snake case to camel case, so it matches the rest of the project.

PR #188, opened 2023-08-05
Use `ESNext` as default module system

jonkoops:
Configures the TypeScript compiler to use `ESNext` as the default module system, this allows the source imports to use the `.js` extension, which is the standard method of importing modules. This PR also cleans up the TSConfigs by removing properties that are already configured, and by placing the configuration for the Rollup build in a separate file.


jest.config.ts
```diff
@@ -11,6 +11,9 @@ const jestConfig: JestConfigWithTsJest = {
     "default",
     ["jest-junit", { outputDirectory: "test-results/jest" }],
   ],
+  moduleNameMapper: {
+    "^(\\.{1,2}/.*)\\.js$": "$1",
+  },
```
jonkoops, line 16: Borrowed this snippet from #170, as I have yet to clean up the Jest config and I don't want to spend too much time on it for this PR.

tsconfig.json
```diff
@@ -1,15 +1,9 @@
...
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": true,
-    "sourceMap": false,
-    "moduleResolution": "node",
-    "noImplicitAny": true,
-    "downlevelIteration": true,
-    "noImplicitThis": true,
+    "target": "ESNext",
+    "module": "ESNext",
+    "moduleResolution": "Node",
```
jonkoops, line 5: We'll have to see if we can change this to `Node16` or `NodeNext`, but since some of our dependencies ship with broken types we cannot do so now.

tsconfig.json
jonkoops, line 1: I've removed most of the configuration in this file as the properties were either not required, or are already set implicitly by setting `strict`.

tsconfig.test.json
```diff
@@ -1,13 +1,6 @@
...
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "noImplicitAny": false,
-    "target": "es6",
-    "noImplicitThis": false,
-    "alwaysStrict": false,
-    "strictBindCallApply": false,
-    "strictNullChecks": false,
-    "strictFunctionTypes": false,
-    "strictPropertyInitialization": false
```
jonkoops, line 11: The tests compile just fine without all these flags disabled, so I've simplified the config by aligning it as much as possible with the root config.

tsconfig.test.json
jonkoops, line 1: The tests compile just fine without all these flags disabled, so I've removed the test config now that it aligns with the base config.

jest.config.ts
jonkoops, line 1: Cleaned up the Jest config so that properties inherited from the `ts-jest` preset are not duplicated.

jest.config.ts
```diff
@@ -1,25 +1,18 @@
 import { type JestConfigWithTsJest } from "ts-jest";
 
 const jestConfig: JestConfigWithTsJest = {
-  rootDir: "./",
-  moduleFileExtensions: ["ts", "js"],
+  preset: "ts-jest/presets/default-esm",
+  moduleNameMapper: {
+    "^(\\.{1,2}/.*)\\.js$": "$1",
+  },
```
jonkoops, line 7: Borrowed this snippet from #170, which matches the recommendations of the [`ts-jest` docs](https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/#use-esm-presets).

frederikprijck (approved):

PR #243, opened 2023-10-30
Add instuctions to README for React Native users

jonkoops:
Adds some instructions to the README specifically for users of React Native. Since the run-time used by React Native, [does not support](https://github.com/facebook/hermes/issues/1178) the global [`atob()`](https://developer.mozilla.org/en-US/docs/Web/API/atob) function needed for `jwt-decode` to function.

For more background information see the discussion under #241


jonkoops:
Looks like I was too slow, closing this in favor of #242

